PR #61, opened 2021-12-01
remove shebang from generated bash and tcsh scripts

amurzeau:
I'm proposing this changes because:
 * Auto completion scripts are not meant to be executable, but only sourced.
 * On Debian systems, bash autocompletion scripts have a warning if they contains a hashbang

See also: https://salsa.debian.org/lintian/lintian/-/merge_requests/292#note_139419


casperdcl (approved):
Thanks!

PR #108, opened 2022-09-09
Fix subcommands with hyphens (and periods, spaces, etc) in zsh

brentyi:
A call to `wordify()` was missing.

Example that breaks with the current version on master:
```python
import argparse

import shtab

parser = argparse.ArgumentParser(prog="test.py")
subparsers = parser.add_subparsers()

sub1 = subparsers.add_parser("subcommand-1", help="First subcommand")
sub2 = subparsers.add_parser("subcommand-2", help="Second subcommand")


print(shtab.complete_zsh(parser))
```


casperdcl (approved):
Thanks! Sorry for the delay

PR #135, opened 2023-04-23
String formatting - f-strings (almost) everywhere

omesser:
... f-strings, f-stringss everywhere

Just a (hopefully...) non-functional PR - all for ✨  of using f-strings almost everywhere, except where it still doesn't make sense and `.format()` is better way of keeping readability.

This can be kinda opinionated, so let's see if maintainers (namely @0x2b3bfa0 and @casperdcl) feels the same way about the choices.


casperdcl (approved):
f-strings make sense as it's a py>=3.7 lib now.

I reverted the other stuff as I prefer the more permissive `yapf` to opinionated `black` style.

PR #143, opened 2023-07-28
bash `nounset` mode fixes

scop:
Fixes issues I came across while briefly testing the shtab completions of `keyring` in `set -o nounset` mode, as well tests failing in that mode.

This might not be a complete set, but it's a start.


casperdcl (approved):
thanks!

PR #144, opened 2023-07-28
Match bash current word exactly instead of as regex

scop:
Matching as regex does not seem correct for these cases to me offhand.


casperdcl (approved):
oops; good spot 🦅👁️ - thanks!

PR #147, opened 2023-07-29
More bash variable leak fixes

scop:


casperdcl (approved):
much local such thanks!

PR #186, opened 2025-05-05
Implement file globbing support

### LINKED ISSUE #69: Function for pattern match file completion

### LINKED ISSUE #124: [feature] Can it provide a function to generate code like <https://github.com/iterative/shtab/blob/main/examples/customcomplete.py#L11-L22>
Can it provide a function to generate code like <https://github.com/iterative/shtab/blob/main/examples/customcomplete.py#L11-L22>? TIA!

simaoafonso-pwt:
This is tested locally in tcsh and bash, zsh is untested.

There should not be any regressions, the test suite passes. We can bikeshed the name.

I did not add support for globbing folders, this doesn't seem to be supported in zsh. I also did not implement the sub-folder support (that is, showing all folders on the file glob completion, so you can descend the hierarchy) since this is only supported in bash. I accept this feature might be required, and are willing to implement it.

---

This is based on https://github.com/iterative/shtab/blob/main/examples/customcomplete.py#L11-L22.

For zsh, https://zsh-manual.netlify.app/completion-system mentions there's no support for globbing folders.

Fix #124 #69

### DIFF
--- a/docs/use.md
+++ b/docs/use.md
@@ -152,11 +152,11 @@ Assuming this code example is installed in `MY_PROG.command.main`, simply run:
     folder for more.
 
 Complex projects with subparsers and custom completions for paths matching
-certain patterns (e.g. `--file=*.txt`) are fully supported (see
+certain patterns (e.g. `--file=*.txt`) are supported using the `fglob` function
+(see fully custom examples
 [examples/customcomplete.py](https://github.com/iterative/shtab/tree/main/examples/customcomplete.py)
 or even
-[iterative/dvc:commands/completion.py](https://github.com/iterative/dvc/blob/main/dvc/commands/completion.py)
-for example).
+[iterative/dvc:commands/completion.py](https://github.com/iterative/dvc/blob/main/dvc/commands/completion.py)).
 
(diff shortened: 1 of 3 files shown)


simaoafonso-pwt:
Rebased to the latest version.

casperdcl (commented):
Thanks for this effort! I just implemented a simpler, more generic (compatible with all shells) solution in #226 - lemme know what you think :)

simaoafonso-pwt:
> Thanks for this effort! I just implemented a simpler, more generic (compatible with all shells) solution in #226 - lemme know what you think :)

That looks like a better solution, looks good to me.

Will close this MR in favor of #226 .

PR #188, opened 2025-05-10
chore: add codespell pre-commit hook

mdeweerd:
- Add codespell pre-commit hook to check for common misspellings
- Configure codespell to skip specific files and directories
- Add codespell configuration in pyproject.toml


mdeweerd:
Note - this was triggered by seeing delimiter misspelled in the generated bash file.

PR #196, opened 2025-10-13
Fix bash quoting, add type hints, code cleanup

pmhahn:
I found your project while looking for how to get bash completion automatically from `ArgumentParser.` Sadly Ubuntu 24.04 ships a broken version, so I had a look at your GitHub repository. Great work. Sadly there are several pending open issues and other PRs, so I hope you will find some time to work on this again soon.

1. While looking at the code I noticed that type-annotations were incomplete. As I'm a huge fan of type-annotated code myself, I took my time to add it to all functions.[^1]
2. I upgraded the old annotations to use `list[…]` instead of `typing.Lits[…]` and such.
3. As usual `mypy` found some minor issues, which I also fixed. See individual commits at the front.
4. While looking at the generated code `shellcheck` found many issues, mostly related to missing quotes. I fixed the code, which generates those. See individual commits in the middle.
5. `__init__.py` is quiet large and contains mixed Python/shell code. Maybe it might be a good idea to move the shell-code to separate files; see last commit.

[^1]: For complete type annotations [`typing.Protocol`](https://docs.python.org/3/library/typing.html#typing.Protocol) from Python 3.8 ([EoL last year](https://devguide.python.org/versions/)) is needed. Therefor I bumped the minimum supported version to 3.8 not that 3.14 got released. By using `from __future__ import annotations`  and `if TYPE_CHECKING:` the code still runs with 3.8. Using `from collections.abc import …` instead of `from typing import …` works only since Python 3.9 (EoL this month). Raising the minimum version to `3.9` thus might be preferred.

If you prefer I can send you smaller pull requests only doing one thing at a time. But before I spend more time on this, I'd like to get your feedback if such changes are preferred or not.

### DIFF
--- a/shtab/__init__.py
+++ b/shtab/__init__.py
@@ -439,18 +439,17 @@ ${root_prefix}() {
 
   if [[ $pos_only = 0 && "${completing_word}" == -* ]]; then
     # optional argument started: use option strings
-    COMPREPLY=( $(compgen -W "${current_option_strings[*]}" -- "${completing_word}") )
+    mapfile -t COMPREPLY < <(compgen -W "${current_option_strings[*]}" -- "${completing_word}")
   elif [[ "${previous_word}" == ">" || "${previous_word}" == ">>" ||
           "${previous_word}" =~ ^[12]">" || "${previous_word}" =~ ^[12]">>" ]]; then
     # handle redirection operators
-    COMPREPLY=( $(compgen -f -- "${completing_word}") )
+    mapfile -t COMPREPLY < <(compgen -f -- "${completing_word}")
   else
     # use choices & compgen
-    local IFS=$'\\n' # items may contain spaces, so delimit using newline
-    COMPREPLY=( $([ -n "${current_action_compgen}" ] \\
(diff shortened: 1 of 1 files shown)


casperdcl (changes requested):
Thanks @pmhahn - apologies for not reviewing sooner.

Would very much prefer solely the `quote` (and any bugfixes) in a separate PR - which I'd be happy to merge; and the rest (stylistic stuff) can be debated in other PR(s).

pmhahn:
> Thanks @pmhahn - apologies for not reviewing sooner.
> 
> Would very much prefer solely the `quote` (and any bugfixes) in a separate PR - which I'd be happy to merge; and the rest (stylistic stuff) can be debated in other PR(s).

Sorry for the delay. I have force-pushed my branch only containing the 2 commits fixing quoting.

How do you want my other changes? I can try to split it into smaller chunks by topic if you prefer.

pmhahn:
@casperdcl Ping?

casperdcl (approved):
forgot to monitor this repo. I wonder if I need to migrate this away from @iterative /CC @0x2b3bfa0 😅

Issue #183, opened 2025-03-11
zsh: `doubled rest argument definition: *::: :->b4` seen in b4


tamird:
This is a weird one, and it is seen with the zsh completions generated from [b4](https://github.com/mricon/b4).

The first time I ever type `b4` in a shell and hit tab I see:
```
tamird@Tamirs-MacBook-Pro b4 % b4
trailers  shazam  send  prep  mbox  diff  ty  pr  kr  am  --
```
however *every* subsequent attempt to complete a b4 command produces:
```
tamird@Tamirs-MacBook-Pro b4 % b4
_arguments:comparguments:327: doubled rest argument definition: *::: :->b4
tamird@Tamirs-MacBook-Pro b4 % b4 pr
_arguments:comparguments:327: doubled rest argument definition: *::: :->b4
```

e3krisztian:
This is the output I got with `shtab==1.7.2`

<details><summary>`bead -s zsh` (click to open)</summary>

    #compdef bead
    
    # AUTOMATICALLY GENERATED by `shtab`
    
    
    _shtab_bead_commands() {
      local _commands=(
        "box:"
        "develop:Unpack a bead as a source tree. Bead directory layout is created, but only the source files are extracted."
        "input:"
        "new:Create and initialize new workspace directory for a new bead."
        "save:Save workspace in a box."
        "status:Show workspace status - name of bead, inputs and their unpack status."
        "version:Show program version info"
        "web:Visualize the big picture."
        "xmeta:eXport eXtended meta attributes to a file next to zip archive."
        "zap:Delete the current workspace directory - like rm -rf \"\$PWD\", only more aggressive."
      )
      _describe 'bead commands' _commands
    }
    
    _shtab_bead_box_commands() {
      local _commands=(
        "add:Define a box."
        "forget:Remove the named box from the boxes known by the tool."
        "list:List boxes."
        "rewire:Remap inputs."
      )
      _describe 'bead box commands' _commands
    }
    
    _shtab_bead_input_commands() {
      local _commands=(
        "add:Make data from another bead available in the input directory."
        "delete:Forget all about an input."
        "load:Put defined input data in place."
        "map:Change the name of the bead from which the input is loaded\/updated."
        "rm:Forget all about an input."
        "unload:Remove input data."
        "update:Update input\[s\] to newest version or defined bead."
      )
      _describe 'bead input commands' _commands
    }
    
    _shtab_bead_options=(
[...]

PR #198, opened 2025-11-16
zsh completion fixes

ThomasWaldmann:
While searching for the root cause of #183, I also had a look into #93 and noticed a discrepancy.

See the related mapping in `format_positional()`, line 515.

I have no clue about zsh scripting or shell completions, but that looked suspicious.

Please review if my fix is actually correct.

### DIFF
--- a/shtab/__init__.py
+++ b/shtab/__init__.py
@@ -606,7 +606,7 @@ def complete_zsh(parser, root_prefix=None, preamble="", choice_functions=None):
 
         return f"""\
 {prefix}() {{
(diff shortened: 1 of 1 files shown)


ThomasWaldmann:
@Freed-Wu @casperdcl please check. ^

shtab/__init__.py
ThomasWaldmann: compare to this!

casperdcl (approved):
🦅👁️ follow-up to/fixes #93

PR #208, opened 2025-11-19
fix typo -> delimiter

ThomasWaldmann:


pmhahn (approved):
Looks good; `uvx codespell` found the same and the same fixes.

casperdcl (approved):

PR #209, opened 2025-11-19
test: capture stdout/stderr in Bash.test helper

ThomasWaldmann:
Previously, the `Bash.test` helper did not capture stdout or stderr,
causing them to be None. This meant that the assertion `not stdout`
was always true (vacuously), and if a test failed, the error message
contained no output.

Changed the assertion and also the assertion failure message to not
hide the "None/None" issue.

This commit adds output capturing to `subprocess.Popen` and decodes
it as utf-8, ensuring that test failures provide useful debugging
information.

### DIFF
--- a/tests/test_shtab.py
+++ b/tests/test_shtab.py
@@ -21,14 +21,19 @@ class Bash:
     def test(self, cmd="1", failure_message=""):
         """Equivalent to `bash -c '{init}; [[ {cmd} ]]'`."""
         init = self.init + "\n" if self.init else ""
-        proc = subprocess.Popen(["bash", "-o", "pipefail", "-euc", f"{init}[[ {cmd} ]]"])
+        proc = subprocess.Popen(
+            ["bash", "-o", "pipefail", "-euc", f"{init}[[ {cmd} ]]"],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            encoding="utf-8",
+        )
         stdout, stderr = proc.communicate()
-        assert (0 == proc.wait() and not stdout and not stderr), f"""\
+        assert (0 == proc.wait() and stdout == "" and stderr == ""), f"""\
 {failure_message}
 {cmd}
 === stdout ===
-{stdout or ""}=== stderr ===
-{stderr or ""}"""
+{stdout}=== stderr ===
+{stderr}"""
 
     def compgen(self, compgen_cmd, word, expected_completions, failure_message=""):
         self.test(


ThomasWaldmann:
Brought to you by google antigravity. :-)

ThomasWaldmann:
> Would prefer if you could repurpose this PR to fix #243 :)

I asked Claude to do it, this is the result. Review carefully! :)

tests/test_shtab.py
```diff
@@ -25,39 +30,191 @@ def complete(parser, shell, **kwargs):
...
+COMP_LINE={shlex.quote(cmdline)}
+COMP_POINT=${{#COMP_LINE}}
+COMP_WORDS=({" ".join(map(shlex.quote, words))})
+COMP_CWORD={cword}
+COMPREPLY=()
+{func}
+printf '%s\\n' "${{COMPREPLY[@]}}\""""
+        proc = subprocess.run(['bash', '-o', 'pipefail', '-uc', completion + driver], cwd=cwd,
+                              capture_output=True, text=True)
+        assert not proc.stderr, proc.stderr
```
yangfan-yf-yf, line 77: This checks only stderr and parses stdout even when Bash exits with a non-zero status. For example, `bad() { exit 7; }; complete -F bad myprog` is reported as an empty candidate list, so a broken completion is indistinguishable from a valid no-match result. The trailing `printf` also masks a completion function that returns non-zero without exiting. Please propagate the function status, check `proc.returncode`, and add regression cases for both `exit` and `return`.

tests/test_shtab.py
```diff
@@ -25,39 +30,191 @@ def complete(parser, shell, **kwargs):
...
+        os.environ['COLUMNS'] = '999'
+        os.execvp(shell, [shell, '-f', '-i'])
+    output = ""
+
+    def read(prompts, timeout=10.0):
+        """Read until the prompt has been seen `prompts` times (or `timeout` elapses)."""
+        nonlocal output
+        strip = r"\x1b\[[0-9;?]*[A-Za-z]|\x1b[=>]|[\a\r\b]"
+        deadline = time.time() + timeout
+        while output.count(PROMPT) < prompts and time.time() < deadline:
```
yangfan-yf-yf, line 119: When the deadline expires, `read()` returns partial output without verifying that the requested prompt was observed. The caller then parses that output as candidates, so an empty result can satisfy negative assertions even though zsh or tcsh never completed setup or completion. Please raise on a missing prompt and apply a bounded deadline to the drain phase as well.

tests/test_shtab.py
```diff
@@ -381,57 +443,59 @@ def test_parser():
...
-def test_fish_global_option_value(test_parser):
+@fix_shell
+def test_global_option_value(shell, change_dir, test_parser):
     """Subcommands complete after `--global-opt value` (#228)"""
-    completion = complete(test_parser, 'fish')
-    candidates = fish_candidates(completion, "myprog --repo x ")
-    assert {"create", "delete", "list"} <= set(candidates)
+    completion = complete(test_parser, shell)
+    if shell == 'tcsh':
+        pytest.xfail("tcsh completes files instead of subcommands after a global option's value")
```
yangfan-yf-yf, line 472: This calls `pytest.xfail()` before the affected candidate query and assertion. The branch therefore never checks the documented tcsh behavior and can never XPASS after it is fixed. The same pattern occurs in the other four shell-specific xfails on this head. Please mark the affected parameter with `pytest.mark.xfail(..., strict=True)` so that the query and assertion still run.

PR #211, opened 2025-12-08
5 small fixes

pmhahn:
This is the next part from my original https://github.com/iterative/shtab/pull/196 containing 4 small fixes from running `ruff` and `mypy`.

### DIFF
--- a/shtab/__init__.py
+++ b/shtab/__init__.py
@@ -245,9 +245,6 @@ def get_bash_commands(root_parser, root_prefix, choice_functions=None):
         options_strings_str = "' '".join(get_option_strings(parser))
         option_strings.append(f"{prefix}_option_strings=('{options_strings_str}')")
         for optional in parser._get_optional_actions():
-            if optional == SUPPRESS:
-                continue
-
             for option_string in optional.option_strings:
                 if hasattr(optional, "complete"):
                     # shtab `.complete = ...` functions
--- a/shtab/__init__.py
+++ b/shtab/__init__.py
@@ -437,8 +437,7 @@ ${root_prefix}() {
   if [[ $pos_only = 0 && "${completing_word}" == -* ]]; then
     # optional argument started: use option strings
     COMPREPLY=( $(compgen -W "${current_option_strings[*]}" -- "${completing_word}") )
-  elif [[ "${previous_word}" == ">" || "${previous_word}" == ">>" ||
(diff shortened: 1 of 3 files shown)


shtab/__init__.py
```diff
@@ -457,22 +434,19 @@ def complete_bash(parser, root_prefix=None, preamble="", choice_functions=None):
...
 
-  COMPREPLY=()
   if [[ $pos_only = 0 && "${completing_word}" == -* ]]; then
     # optional argument started: use option strings
-    while IFS= read -r line; do COMPREPLY+=("$line"); done < <(
-      compgen -W "${current_option_strings[*]}" -- "${completing_word}")
-  elif [[ "${previous_word}" == ">" || "${previous_word}" == ">>" ||
-          "${previous_word}" =~ ^[12]">" || "${previous_word}" =~ ^[12]">>" ]]; then
+    COMPREPLY=( $(compgen -W "${current_option_strings[*]}" -- "${completing_word}") )
+  elif [[ "${previous_word}" =~ ^[0-9\\&]*[\\<\\>]\\>?$ ]]; then
```
casperdcl, line 465: what's the reasoning for this?

pmhahn:
> I think there are 2 changes still outstanding in this PR (the others are already addressed by various other PRs).
> 
> why remove this?
> 
> ```python
> if optional == SUPPRESS:
>     continue
> ```

Quoting the commit message of that commit:
> ArgumentParser._get_optional_actions() will always return instances of
>  `Action` and never `SUPPRESS`, which is a `str`.

Running any Python type-checker like `mypy`, `pyritgh`, `pyrefly`, `ty` will tell youm that `self._actions: list[_ActionT]` will never match `SUPPRESS: str`, so that `if` will never evaluate to `true` and should be removed (or fixed).

shtab/__init__.py
pmhahn: Quoting from the commit message of that commit:
> Fix detection of redirection
>
> Also handle input redirection.
> Also handle other file descriptors than STDOUT and STDERR.

file-descriptors are just numbers and there are (many) more than just 1 and 2 for `STDOUT` and `STDERR`; `0` for `STDIN` but any other (non-negative) number is valid; and there are more variants than only _overwriting_ with `>` and _appending_ with `>>`, e.g. `<` for input-redirection, `<>` for read-write, …

casperdcl: ah I see, basically `&` and other `fd` numbers added.

Issue #199, opened 2025-11-16
split shtab/__init__.py?


ThomasWaldmann:
Maybe it would be a good idea to split that:
- separate python code files, 1 per shell (currently: bash, zsh, tcsh)
- common code used by multiple shells either stays in `__init__.py` or is moved to `common.py`

First merge PRs that add support for a new shell to avoid merge conflicts.

ThomasWaldmann:
Note: #196 has been merged, but it is still everything in `__init__.py`.

ThomasWaldmann:
BTW, it might be an interesting experiment to let claude dissect this.

@casperdcl What do you think, are there enough tests to make sure we would notice any potential damage?

casperdcl:
please not AI 😅

ThomasWaldmann:
OK.

PR #212, opened 2026-02-14
Add PowerShell tab completion support

HeyItsGilbert:
## Summary

This PR adds PowerShell tab completion support to shtab, enabling native PowerShell argument completion for CLI applications built with argparse.

**Key features:**
- Generates PowerShell completion scripts using `Register-ArgumentCompleter -Native`
- Supports PowerShell 5.1+ (Windows Desktop) and PowerShell Core 7.x (cross-platform)
- Automatically discovers subcommands, options, choices, and file/directory completions
- Follows the same architecture as existing bash/zsh/tcsh support
- Includes examples and comprehensive documentation

## Changes

- **shtab/__init__.py**: Added PowerShell completer implementation with:
  - `get_powershell_commands()` for parser tree extraction
  - `complete_powershell()` decorated completer function
  - PowerShell script template with state machine for completion dispatch
  - Serialization helpers for PowerShell hashtable generation
  - Integration with existing `CHOICE_FUNCTIONS` infrastructure

- **Documentation & Examples**:
  - Updated `README.rst` and `docs/index.md` to list PowerShell as supported shell
  - Added PowerShell installation & usage guide in `docs/use.md`
  - Updated `examples/customcomplete.py` with PowerShell preamble

- **Tests**: Extended existing parametrized test suite to include PowerShell

### Prerequisites
```bash
pip install -e .
```

### Run tests
```bash
# All tests (including PowerShell smoke tests)
pytest tests/test_shtab.py -v

# PowerShell-only tests (requires pwsh)
pytest tests/test_shtab.py -k "powershell" -v
```

### Generate completion script manually
```bash
# For shtab itself
shtab --shell=powershell shtab > shtab.ps1

# Or for an example app
python -c "from examples.customcomplete import main; print(main)" | shtab --shell=powershell > custom.ps1
```

### Test completion in PowerShell
```powershell
# Load the generated script
. ./shtab.ps1

# Test completions (Ctrl+Space or type -<TAB>)
shtab -<TAB>
```

### Enable debug output
Set the `SHTAB_DEBUG` environment variable to trace completion state:
```powershell
$env:SHTAB_DEBUG=1
. ./shtab.ps1
```

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -18,6 +18,7 @@
     - `bash`
     - `zsh`
     - `tcsh`
+    - `powershell`
 -   Supports
     - [argparse](https://docs.python.org/library/argparse)
     - [docopt](https://pypi.org/project/docopt) (via [argopt](https://pypi.org/project/argopt))
@@ -110,7 +111,6 @@ application. Use `-u, --error-unimportable` to noisily complain.
 Please do open [issues][GH-issue] & [pull requests][GH-pr]! Some ideas:
 
 - support `fish` ([#174](https://github.com/iterative/shtab/pull/174))
-- support `powershell`
 
 See
 [CONTRIBUTING.md](https://github.com/iterative/shtab/tree/main/CONTRIBUTING.md)
--- a/docs/use.md
+++ b/docs/use.md
@@ -7,8 +7,8 @@ There are two ways of using `shtab`:
     - end-users execute `shtab your_cli_app.your_parser_object`
 - [Library Usage](#library-usage): as a library integrated into your CLI application
     - adds a couple of lines to your application
-    - argument mode: end-users execute `your_cli_app --print-completion {bash,zsh,tcsh}`
-    - subparser mode: end-users execute `your_cli_app completion {bash,zsh,tcsh}`
+    - argument mode: end-users execute `your_cli_app --print-completion {bash,zsh,tcsh,powershell}`
+    - subparser mode: end-users execute `your_cli_app completion {bash,zsh,tcsh,powershell}`
 
 ## CLI Usage
 
@@ -98,6 +98,32 @@ Below are various examples of enabling `shtab`'s own tab completion scripts.
       | sudo tee /etc/profile.d/eager-completion.csh
     ```
 
+=== "powershell"
+
+    ```powershell
+    shtab --shell=powershell shtab.main.get_main_parser --error-unimportable `
+      | Out-File -FilePath ~\shtab_completion.ps1
+    . ~\shtab_completion.ps1
+    ```
+
+=== "Eager powershell"
+
+    Add the following to your PowerShell profile (`$PROFILE`):
+
+    ```powershell
+    shtab --shell=powershell shtab.main.get_main_parser `
+      | Out-String | Invoke-Expression
+    ```
+
+    Or save to a file and dot-source it from your profile:
+
+    ```powershell
+    shtab --shell=powershell shtab.main.get_main_parser `
+      | Out-File -FilePath ~\shtab_completion.ps1
+    # Add to $PROFILE:
+    . ~\shtab_completion.ps1
+    ```
+
 !!! tip
     See the [examples/](https://github.com/iterative/shtab/tree/main/examples)
     folder for more.
@@ -143,6 +169,14 @@ Assuming this code example is installed in `MY_PROG.command.main`, simply run:
     ```sh
     shtab --shell=tcsh -u MY_PROG.command.main.get_main_parser \
       | sudo tee /etc/profile.d/MY_PROG.completion.csh
+    ```
+
+=== "powershell"
+
+    ```powershell
+    shtab --shell=powershell -u MY_PROG.command.main.get_main_parser `
+      | Out-File -FilePath ~\MY_PROG_completion.ps1
+    . ~\MY_PROG_completion.ps1
     ```
 
 ## Library Usage
--- a/shtab/__init__.py
+++ b/shtab/__init__.py
@@ -38,10 +38,11 @@ log = logging.getLogger(__name__)
 SUPPORTED_SHELLS: List[str] = []
 _SUPPORTED_COMPLETERS = {}
 CHOICE_FUNCTIONS: Dict[str, Dict[str, str]] = {
-    "file": {"bash": "_shtab_compgen_files", "zsh": "_files", "tcsh": "f",
-             "powershell": "_shtab_powershell_compgen_files"},
-    "directory": {"bash": "_shtab_compgen_dirs", "zsh": "_files -/", "tcsh": "d",
-                  "powershell": "_shtab_powershell_compgen_dirs"}}
+    "file": {
+        "bash": "_shtab_compgen_files", "zsh": "_files", "tcsh": "f",
+        "powershell": "_shtab_powershell_compgen_files"}, "directory": {
+            "bash": "_shtab_compgen_dirs", "zsh": "_files -/", "tcsh": "d",
+            "powershell": "_shtab_powershell_compgen_dirs"}}
 FILE = CHOICE_FUNCTIONS["file"]
 DIRECTORY = DIR = CHOICE_FUNCTIONS["directory"]
 FLAG_OPTION = (
@@ -920,7 +921,8 @@ def get_powershell_commands(root_parser, root_prefix, choice_functions=None):
                 opt_key = f"{prefix}_{wordify(option_string)}"
 
                 if hasattr(optional, "complete"):
-                    comp_pattern = complete2pattern(optional.complete, "powershell", choice_type2fn)
+                    comp_pattern = complete2pattern(optional.complete, "powershell",
+                                                    choice_type2fn)
                     if comp_pattern:
                         all_compgens[opt_key] = comp_pattern
 
--- a/shtab/__init__.py
+++ b/shtab/__init__.py
@@ -38,8 +38,10 @@ log = logging.getLogger(__name__)
 SUPPORTED_SHELLS: List[str] = []
 _SUPPORTED_COMPLETERS = {}
 CHOICE_FUNCTIONS: Dict[str, Dict[str, str]] = {
-    "file": {"bash": "_shtab_compgen_files", "zsh": "_files", "tcsh": "f"},
-    "directory": {"bash": "_shtab_compgen_dirs", "zsh": "_files -/", "tcsh": "d"}}
+    "file": {"bash": "_shtab_compgen_files", "zsh": "_files", "tcsh": "f",
+             "powershell": "_shtab_powershell_compgen_files"},
+    "directory": {"bash": "_shtab_compgen_dirs", "zsh": "_files -/", "tcsh": "d",
+                  "powershell": "_shtab_powershell_compgen_dirs"}}
 FILE = CHOICE_FUNCTIONS["file"]
 DIRECTORY = DIR = CHOICE_FUNCTIONS["directory"]
 FLAG_OPTION = (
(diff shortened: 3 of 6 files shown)


shtab/__init__.py
```diff
@@ -38,8 +38,11 @@
 SUPPORTED_SHELLS: List[str] = []
 _SUPPORTED_COMPLETERS = {}
 CHOICE_FUNCTIONS: Dict[str, Dict[str, str]] = {
-    "file": {"bash": "_shtab_compgen_files", "zsh": "_files", "tcsh": "f"},
-    "directory": {"bash": "_shtab_compgen_dirs", "zsh": "_files -/", "tcsh": "d"}}
+    "file": {
+        "bash": "_shtab_compgen_files", "zsh": "_files", "tcsh": "f",
+        "powershell": "_shtab_powershell_compgen_files"}, "directory": {
+            "bash": "_shtab_compgen_dirs", "zsh": "_files -/", "tcsh": "d",
+            "powershell": "_shtab_powershell_compgen_dirs"}}
```
pmhahn: Please add line breaks here as you're mixing `file` and `directory` here::

```suggestion
    "file": {
        "bash": "_shtab_compgen_files", "zsh": "_files", "tcsh": "f",
        "powershell": "_shtab_powershell_compgen_files"},
    "directory": {
        "bash": "_shtab_compgen_dirs", "zsh": "_files -/", "tcsh": "d",
        "powershell": "_shtab_powershell_compgen_dirs"},
}
```

shtab/__init__.py
```diff
@@ -800,6 +803,337 @@ def recurse_parser(cparser, positional_idx, requirements=None):
...
 
+def _powershell_escape(string: str) -> str:
+    """Escape a string for use inside a PowerShell single-quoted string."""
+    return str(string).replace("'", "''")
+
+
+def _powershell_list(items):
+    """Serialize a list of strings to a PowerShell array literal."""
+    if not items:
+        return "@()"
```
pmhahn: This `if` block is not needed as `",".join([])` evaluates to the empty string.

```suggestion
```

shtab/__init__.py
```diff
@@ -800,6 +803,337 @@ def recurse_parser(cparser, positional_idx, requirements=None):
...
+    return f"@({escaped})"
+
+
+def _powershell_hashtable(d):
+    """Serialize a dict[str, list[str]] to PowerShell @{} syntax."""
+    if not d:
+        return "@{}"
+    entries = []
+    for key, values in sorted(d.items()):
+        entries.append(f"    '{_powershell_escape(key)}' = {_powershell_list(values)}")
```
pmhahn: List-comprehension please:

```suggestion
    entries = [
        f"    '{_powershell_escape(key)}' = {_powershell_list(values)}"
        for key, values in sorted(d.items())
    ]
```

shtab/__init__.py
```diff
@@ -800,6 +803,337 @@ def recurse_parser(cparser, positional_idx, requirements=None):
...
+    return "@{\n" + "\n".join(entries) + "\n}"
+
+
+def _powershell_flat_hashtable(d):
+    """Serialize a dict[str, str] to PowerShell @{} syntax."""
+    if not d:
+        return "@{}"
+    entries = []
+    for key, value in sorted(d.items()):
+        entries.append(f"    '{_powershell_escape(key)}' = '{_powershell_escape(value)}'")
```
pmhahn: List-comprehension please:

```suggestion
    entries = [
        f"    '{_powershell_escape(key)}' = '{_powershell_escape(value)}'"
        for key, value in sorted(d.items())
    ]
```

shtab/__init__.py
```diff
@@ -800,6 +803,337 @@ def recurse_parser(cparser, positional_idx, requirements=None):
         optionals_special_str=' \\\n        '.join(specials))
 
 
+def _powershell_escape(string: str) -> str:
+    """Escape a string for use inside a PowerShell single-quoted string."""
```
pmhahn: Nit: I would prefer that `_ps1_escape` adds the surrounding single-quotes itself. I'm no PS1 expert myself, but at least with `bash` there are several ways to escape strings
- in single-quotes only single-quotes must be handles special
- in double-quotes several characters still need backslash escaping
- without quote many characters must be backslash escaped

With bash `shlex.quote()` could switch between those choices or even depend on the string to escape and we don't care which is uses as it is an internal details.

With your `_ps1_escape` here you always must use single quotes and we later can't change it if we find something better.

shtab/__init__.py
```diff
@@ -800,6 +803,337 @@ def recurse_parser(cparser, positional_idx, requirements=None):
...
+            [],
+        )
+
+    def recurse(parser, prefix):
+        discovered_subparsers = []
+
+        for i, positional in enumerate(parser._get_positional_actions()):
+            if positional.help == SUPPRESS:
+                continue
+
```
pmhahn: Following 9b2a478acdfc3a0d858051c2ad342ee8440bdbbb this is not necesarry:

```suggestion
```

shtab/__init__.py
```diff
@@ -800,6 +803,337 @@ def recurse_parser(cparser, positional_idx, requirements=None):
...
+    """Serialize a dict[str, str] to PowerShell @{} syntax."""
+    if not d:
+        return "@{}"
+    entries = []
+    for key, value in sorted(d.items()):
+        entries.append(f"    '{_powershell_escape(key)}' = '{_powershell_escape(value)}'")
+    return "@{\n" + "\n".join(entries) + "\n}"
+
+
+def get_powershell_commands(root_parser, root_prefix, choice_functions=None):
```
pmhahn, line 847: Oh oh, I see lots of `get_bash_commands()` code here. There are many differences, but I don't have time to dig in right now to find out, why they differ.

Maybe someone™ should refactor this and to extract common code? 🤔

pmhahn (changes requested):
I just had a short look. Looks okay mostly, but someone with PowerShell knowledge should have a look at the embedded PS1 code.

Found some minor nits.

Maybe split the large `__init__.py` into individual files in the future? Or at least move the embedded shell-scripts parts into individual files? See https://github.com/iterative/shtab/issues/199 ?

shtab/__init__.py
HeyItsGilbert: Do you need this addressed in this PR?

pmhahn: No, but a mental note to whoever to address this soon.

shtab/__init__.py
HeyItsGilbert: I rewrote this a bit to always wrap in single quotes. I did find an very weird edge case, but that should be dealt with now. I checked where I use it, and honestly I don't see where we would want anything but single quoted strings. What the tool passes in should probably never be interpolated. If there is something dynamic, we should do that in the python code to keep things deterministic. If you can think of case, let me know.

shtab/__init__.py
```diff
@@ -800,6 +804,341 @@ def recurse_parser(cparser, positional_idx, requirements=None):
...
+    Quote a string for PowerShell (single-quoted, like ``shlex.quote``).
+    In single-quoted strings the only special characters are quote chars.
+    PowerShell treats smart/curly single quotes (\u2018, \u2019) the same as
+    regular single quotes, so they must be doubled too.
+    See https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_quoting_rules
+    """
+    s = str(string)
+    for ch in ("'", "\u2018", "\u2019"):
+        s = s.replace(ch, ch * 2)
+    return "'" + s + "'"
```
pmhahn: As `str` are immutable in Python, that `s = str(string)` looks strange. Maybe just rename the argument.
And use a f-string to concatenate the strings.

```suggestion
def _powershell_escape(s: str) -> str:
    """
    Quote a string for PowerShell (single-quoted, like ``shlex.quote``).
    In single-quoted strings the only special characters are quote chars.
    PowerShell treats smart/curly single quotes (\u2018, \u2019) the same as
    regular single quotes, so they must be doubled too.
    See https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_quoting_rules
    """
    for ch in ("'", "\u2018", "\u2019"):
        s = s.replace(ch, ch * 2)
    return f"'{s}'"
```

pmhahn (approved):
Sorry for the delay; only 1 nit you might want to fix.

Again: I'm a Linux-only user and have no experience with PS, so I cannot say if the completion works as expected. Nice work, thanks.

shtab/__init__.py
```diff
@@ -71,6 +73,15 @@ def glob(*patterns: str) -> CompleteType:
   end
   __fish_complete_path "$comp" | string match -e "*/"  # recurse into subdirs
 end
+""", "powershell": f"""
+function _shtab_pattern_compgen_{abs(hash(patterns))} {{
+  param([string]$WordToComplete)
+  Get-ChildItem -Path "$WordToComplete*" `
+  -Include '{"','".join(map(_powershell_escape, patterns))}' -File -ErrorAction SilentlyContinue |
+    ForEach-Object {{ $_.Name }}
```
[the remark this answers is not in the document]
casperdcl: should be `_powershell_list` instead

casperdcl (changes requested):
Thanks for this start! Just rebased.

Locally, some things don't work for me:

- [x] `customcomplete -<TAB>` doesn't complete anything
  + document known-issue: works on `pwsh~=7`, not on `powershell.exe~=5`
- [x] `customcomplete process foo csv <tab>` shouldn't complete files
- [x] `customcomplete process <tab>` needs changing `shtab.cmd("head -c5 /dev/random ...")` to something windows-compatible

casperdcl:
Almost there! `-*, --*` still doesn't work:

```sh
customcomplete --<TAB>  # doesn't work
customcomplete --h<TAB> # works
```

HeyItsGilbert:
The new tests should include the things you noted @casperdcl 🤞

shtab/__init__.py
casperdcl, line 1: idk about the test just added, but it doesn't look fixed:

```sh
pathcomplete --<TAB>       # doesn't work
pathcomplete --h<TAB>      # works (--help)
```

also:

```sh
pathcomplete Vi<TAB>       # works (Videos\)
pathcomplete Videos\Ca<TAB># works (Videos\Captures\)
pathcomplete Videos\<TAB>  # doesn't work (Captures\) i.e. strips prefix
```

MohammedAlkindi:
Tested this on real Windows, since PowerShell eyes were asked for earlier in the thread. Setup: PR head `dcfe7cb`, Windows 11 Home, Windows PowerShell 5.1.26100, completions driven programmatically through `TabExpansion2` (that's what PSReadLine calls underneath) with the generated script dot-sourced and a `pathcomplete` stub on `PATH`. Parser is the docs' pathcomplete example (`file` positional = `shtab.FILE`, `--dir` = `shtab.DIRECTORY`), run in a directory containing `Videos\Camera\` and `Videos\Captures\`:

```
INPUT <pathcomplete ->         -> []
INPUT <pathcomplete -->        -> []
INPUT <pathcomplete --h>       -> [--help]
INPUT <pathcomplete Vi>        -> [Videos\]
INPUT <pathcomplete Videos\>   -> [Camera\ | Captures\]        # prefix stripped
INPUT <pathcomplete Videos\Ca> -> [Videos\Camera\ | Videos\Captures\]
```

That reproduces both of @casperdcl's reports exactly, and the two have different causes.

**1. `-<TAB>` / `--<TAB>`: the engine never calls the completer, so this script can't fix it.** I replaced the registered scriptblock with one that unconditionally logs to stderr and returns a dummy result:

```
INPUT <pathcomplete ->   -> []          (no CALLED line — scriptblock never ran)
INPUT <pathcomplete -->  -> []          (no CALLED line)
CALLED word='--h'
INPUT <pathcomplete --h> -> [DUMMY]
```

On PowerShell 5.1, `Register-ArgumentCompleter -Native` is simply not invoked while the word is a bare `-` or `--`; it fires from `--h` onward. Every native completer on that engine behaves this way, so it isn't a defect in the generated script and no test added here can make it pass. I don't have pwsh 7 on this box, so I can't say whether 7.x behaves better — worth one check there before treating this item as a blocker.

**2. `Videos\<TAB>` prefix-strip: real bug in the generated compgen helpers, small fix.** `_shtab_powershell_compgen_files`/`_dirs` derive the prefix to re-attach with `Split-Path -Parent`, and:

```powershell
PS> Split-Path 'Videos\' -Parent
        # '' — for a word ending in the separator, the parent is empty
PS> Split-Path 'Videos\Ca' -Parent
Videos
```

So exactly when the word ends in `\`, results are emitted with no directory prefix, and PowerShell replaces the whole word (`ReplacementLength` covers it) with the bare child name — `Videos\` becomes `Captures\`. `Join-Path` also rewrites whatever prefix the user actually typed: `.\Vi<TAB>` came back as `Videos\` with the `.\` dropped. Keeping the typed prefix verbatim fixes both:

```powershell
  $dir = ''
  if ($WordToComplete -match '^(.*[\\/])') { $dir = $Matches[1] }
  Get-ChildItem -Path "$WordToComplete*" -File -ErrorAction SilentlyContinue |
    ForEach-Object { $dir + $_.Name }
  Get-ChildItem -Path "$WordToComplete*" -Directory -ErrorAction SilentlyContinue |
    ForEach-Object { $dir + $_.Name + [System.IO.Path]::DirectorySeparatorChar }
```

Same harness with that applied to the generated script:

```
INPUT <pathcomplete Videos\>   -> [Videos\Camera\ | Videos\Captures\]
INPUT <pathcomplete Videos\Ca> -> [Videos\Camera\ | Videos\Captures\]
'.\Vi'      -> [.\Videos\]
'Videos/'   -> [Videos/Camera\ | Videos/Captures\]
```

(`_shtab_powershell_join_prefix` has no remaining callers after this.)

Not tested here: interactive PSReadLine rendering (all of the above is programmatic `TabExpansion2`) and pwsh 7.x, which isn't installed on this machine.

shtab/__init__.py
HeyItsGilbert: Are you saying you're not sure about the test because the `--<TAB>` didn't work. Are you testing in `powershell.exe` or `pwsh`? I've been working on this on my debian box (`pwsh`), but I'll pull it to my windows laptop. Mohammed does call out a few known issues with  `Register-ArgumentCompleter -Native` not being called in `powershell.exe`.

HeyItsGilbert:
Here's me testing the next commit on debian and Windows (pwsh).
Debian (pwsh): https://asciinema.org/a/46cx8Y2YGVlEgBVf
Windows (pwsh): 
image

Windows (powershell )
image

shtab/__init__.py
casperdcl: I've been testing on windows native `powershell.exe`. I had no idea there's a `pwsh`. MS continues its identity crisis, I suppose 😅

MohammedAlkindi: `powershell.exe` 5.1 here, not `pwsh`. The two failures have different causes.

`--<TAB>` is not this PR. In Windows PowerShell 5.1 the native completer is never invoked when
the token being completed is a bare `-`, `--`, or a short option like `-s`. I registered a
throwaway completer that unconditionally returns one sentinel result: it also never fires for
those three, and does fire for `` , `--h` and `x`. Nothing in the generated script can reach it.

`Videos\<TAB>` is this PR, and it is two characters. In `_shtab_powershell_compgen_files` and
`_shtab_powershell_compgen_dirs`, the prefix regex is `'^(.*[\/])'`, a class holding only forward
slash, so `$dir` stays empty for Windows separators. `Videos\Ca<TAB>` is broken the same way.

<details>
<summary>Measurements</summary>

Driven through `TabExpansion2`, which returns exactly what TAB would insert.

Native completer reached at all (sentinel completer, same session):

| input | completer ran |
|---|---|
| `pathcomplete ` | yes |
| `pathcomplete x` | yes |
| `pathcomplete --h` | yes |
| `pathcomplete -` | no |
| `pathcomplete --` | no |
| `pathcomplete -s` | no |

`--` and `-s` parse as `CommandParameterAst`; `--h` and `x` parse as `StringConstantExpressionAst`.

Path cases, at `9e58641` and after changing the two regex sites to `[\\/]`:

| input | at 9e58641 | after |
|---|---|---|
| `Videos\` | `pathcomplete note.txt` | `pathcomplete Videos\note.txt` |
| `Videos\Ca` | `pathcomplete Captures\` | `pathcomplete Videos\Captures\` |
| `Videos/` | `pathcomplete Videos/note.txt` | unchanged |
| `Vi` | `pathcomplete Videos\` | unchanged |
| `--h` | `pathcomplete --help` | unchanged |
| `--` | no completions | no completions |

In `shtab/__init__.py` the two sites are lines 1277 and 1287, written `[\\/]`. The template is a
plain `Template("""...` at line 1260, so `\\` collapses to `\` and the emitted class is `[\/]`.
Line 99 already uses `[\\\\/]` and emits `[\\/]` correctly. Fix is those two lines to `[\\\\/]`.

Suite on this box, before and after that change: `18 failed, 54 passed, 68 skipped`, identical.
All 18 are the `[bash]` tests, failing on `execvpe(/bin/bash) failed`, no POSIX bash here.

</details>

docs/index.md
```diff
@@ -41,6 +41,7 @@ shell | location
 [`zsh`](https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org) | `/usr/local/share/zsh/site-functions/_NAME`
 [`tcsh`](https://github.com/tcsh-org/tcsh/blob/master/complete.tcsh) | `/etc/profile.d/completion_NAME.csh`, source in `~/.cshrc` or `~/.tcshrc`
 [`fish`](https://fishshell.com/docs/current/completions.html#where-to-put-completions) | `~/.config/fish/completions/NAME.fish`
+[`powershell`](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/register-argumentcompleter) | `~\.config\powershell\completions\NAME.ps1`, source in `$PROFILE`
```
casperdcl: open to better suggestions for this URL... perhaps https://learn.microsoft.com/en-us/powershell/scripting/learn/shell/creating-profiles?

docs/use.md
```diff
@@ -101,6 +101,32 @@ Below are various examples of enabling `shtab`'s own tab completion scripts.
     # Install system-wide (pkg-config fish --variable=completionsdir)
     shtab -u --shell=fish shtab.main.get_main_parser \
       | sudo tee /usr/share/fish/vendor_completions.d/shtab.fish
+
+=== "powershell"
+
+    ```powershell
+    shtab --shell=powershell shtab.main.get_main_parser --error-unimportable `
+      | Out-File -FilePath ~\.config\powershell\completions\shtab.ps1
```
casperdcl, line 115: open to better suggestions for output location

shtab/__init__.py
```diff
@@ -1088,11 +1111,365 @@ def recurse_parser(cparser: ArgumentParser, path: list[str]):
...
+    #   microsoft.powershell.core/register-argumentcompleter
+    # - https://learn.microsoft.com/en-us/powershell/scripting/
+    #   learn/shell/tab-completion
+    preamble = "\n".join(list(dict.fromkeys(([preamble] if preamble else []) + extra_preambles)))
+    return Template(r"""# AUTOMATICALLY GENERATED by https://github.com/tqdm/shtab
+# Usage:
+# 1) Copy this to somewhere (e.g. ~\.config\powershell\completions\${prog}.ps1
+# 2) Add the following line to your $PROFILE:
+#    . ~\.config\powershell\completions\${prog}.ps1
+
```
casperdcl, line 1265: maybe this?

```suggestion
# See also: https://learn.microsoft.com/en-us/powershell/scripting/learn/shell/creating-profiles

```

docs/use.md
```diff
@@ -7,8 +7,8 @@ There are two ways of using `shtab`:
     - end-users execute `shtab your_cli_app.your_parser_object`
 - [Library Usage](#library-usage): as a library integrated into your CLI application
     - adds a couple of lines to your application
-    - argument mode: end-users execute `your_cli_app --print-completion {bash,zsh,tcsh}`
-    - subparser mode: end-users execute `your_cli_app completion {bash,zsh,tcsh}`
+    - argument mode: end-users execute `your_cli_app --print-completion {bash,zsh,tcsh,fish,powershell}`
+    - subparser mode: end-users execute `your_cli_app completion {bash,zsh,tcsh,fish,powershell}`
```
casperdcl, line 11: I wonder if we should use `pwsh` (or `ps1`) in lieu of `powershell` everywhere?

casperdcl (approved):
just some final docs-related questions.

docs/use.md
MohammedAlkindi: Your regex fix works. On `powershell.exe` 5.1, `Videos\<TAB>` now completes to `Videos\note.txt` and `Videos\Ca<TAB>` to `Videos\Captures\`, with the forward-slash and no-separator cases unchanged.

On the naming: `pwsh` is the more honest label. `--` and `-s` still return nothing on 5.1, because the completer is never invoked for those tokens at all, and that is not reachable from the generated script.

The suite already only covers `pwsh`: `powershell_candidates` skips when it is absent, and `test_powershell_single_token` asserts `myprog --` gives `["--help", "--repo"]`, which cannot hold on 5.1. Renaming would line the label up with what is actually exercised.

HeyItsGilbert: PowerShell is the name of the language. Calling it `pwsh` might decrease discoverability. I do think that calling out that 5.1 (aka `Windows PowerShell` or `powershell.exe`) is only partially supported is a good idea. Frankly anything that nudges folks to PowerShell (aka `pwsh`) is a good thing as that's where MSFT is putting all their effort.

docs/use.md
HeyItsGilbert: That's where I put mine  ¯\\_(ツ)_/¯. It's a valid XDG_CONFIG_HOME path and it still works on Windows. The only thing would be that `.config\powershell\completions` path probably won't exist. We could include the `New-Item -Path "~\.config\powershell\completions" -ItemType Directory -Force` to the docs.

In the next version they're going to release a new cmdlet that let's you determine where you "PowerShell files" are but I wouldn't put that in the docs since you'd have to be on latest.

docs/index.md
HeyItsGilbert: That's probably fine if the goal is to tell users how to setup their `$PROFILE`.

docs/use.md
MohammedAlkindi: Agreed on discoverability, keep `powershell`.

Worth making the 5.1 caveat specific rather than "partial", though. Re-checked just now at `c24a64b`: path completion and long options work (`Videos\<TAB>`, and `--h<TAB>` gives `--help`), but the completer is never invoked at all for a bare `-`, `--`, or a short option like `-s`, so those three silently produce nothing.

casperdcl (approved):
Looks ready to me 🥂

Would be nice to speedup the tests (and perhaps include `pwsh` in the `ubuntu` CI, and consider dropping macos & windows CI) but that can be in another PR.

Please do open follow-ups if you spot anything!

PR #213, opened 2026-03-23
improvements and fixes to the tcsh completion logic

ThomasWaldmann:
Fixes and completes tcsh support for subcommand positionals, and adds `--opt=<TAB>` completion.

Rebased onto current `main`, so this now supersedes the `p@N@`-rule generation added in
`d9b378598` ("support positional choices") rather than conflicting with it.

### 1. Positional completion under subcommands, at any depth

Sub-parsers are recursed into regardless of how many requirements were already collected (before:
only at the top level, `if not requirements`), and each branch becomes a conditional:

```
if ( $#cmd >= 3 && ("$cmd[2]" == "key") && ("$cmd[3]" == "change-location") ) echo repokey keyfile
```

so `prog key change-location <TAB>` completes, not just `prog key <TAB>`.

### 2. No more out-of-range array indexing

tcsh evaluates `$cmd[2]` even when `$cmd` is shorter, which surfaces as `if: Empty if.` or
`Subscript out of range`. Guarded with an explicit `$#cmd >= max_idx` check plus empty-string
padding of `$cmd`.

`$COMMAND_LINE` has to stay **unquoted** in that assignment. Quoting it stops csh word splitting,
so `$cmd[1]` becomes the whole command line and `$cmd[2]` is always empty — no subcommand-level
completion would ever fire.

### 3. Arguments with a custom `.complete` in the multi-requirement branch

Previously only `arg.choices` was handled there, so such arguments produced an empty rule body
(e.g. `'p@3@`set cmd=($COMMAND_LINE); `@'`). Nested backticks crash tcsh's parser, so a backticked
pattern is invoked via `eval` instead of being nested inside the outer backticks.

### 4. `--opt=<TAB>` completion for options taking a value (`nargs != 0`)

Emitted as `c/--opt=/` rules — **before** the generic `c/--/(...)` rule. tcsh uses the first rule
whose *pattern* matches, even if that rule then yields no candidates, so in the other order
`--opt=<TAB>` matched `c/--/` , tried to complete `opt=…` against the list of long option names,
found nothing, and stopped. Ordering them first is safe: `--<TAB>` does not match `c/--opt=/` and
still falls through to the generic rule.

### 5. Rule deduplication

`specials` is deduplicated, preserving order. tcsh silently truncates oversized completion
definitions, which shows up as cryptic `Illegal variable name.` errors.

---

Verified by driving a real interactive tcsh 6.21.00 with a parser that has nested subcommands, an
argument with a custom `.complete`, and a valued option:

| completion | before | after |
| --- | --- | --- |
| `demo key <TAB>` | ✅ | ✅ |
| `demo key change-location <TAB>` | ❌ nothing | ✅ `keyfile repokey` |
| `demo key export <TAB>` (custom `.complete`) | ❌ nothing | ✅ `alpha beta gamma` |
| `demo --out=b<TAB>` | ❌ nothing | ✅ files |
| `demo --<TAB>` | ✅ | ✅ (unchanged) |

The existing test suite is unaffected (one bash test fails for me both with and without this
branch — bash 3.2 on macOS — so it looks environmental).

Not covered here: `tcsh` still has no equivalent of the archive/`aid:`/tags dynamic completions.


ThomasWaldmann:
Note:
- code change was AI generated (as a side effect of working on borgbackup's tcsh completions, where the AI noticed some troubles while testing these changes)
- I transformed the original monkeypatch to a minimal patch for shtab, removed cosmetic changes like python quotes, line wrapping, etc.
- I did a quick test of the patched shtab code with borgbackup, it now works for subcommands.

This needs review by someone with a clue about tcsh and shtab (that's not me).

ThomasWaldmann:
Maybe @simaoafonso-pwt could help here.

I set "allow edits by maintainers" - @simaoafonso-pwt if you are not maintainer here, feel free to copy whatever you like from here to an own PR.

simaoafonso-pwt:
> It now adds completions for optional arguments that take parameters (nargs != 0) by appending a = version (e.g., --opt=).

This seems like a good feature, but it doesn't really work. I see it produces:

```
        'c/-/(- h o v)/' \
...
        'n/-o/f/' \
        'c/-o=/f/' \
```

The `-o=` needs to be include before `c/-/...`, otherwise it will not reach it.

---

For the rest of the changes, I don't have so many complex commands to test, if you say that borg works better, I'm good.

ThomasWaldmann:
Updated (force-push `4142cdd` → `8d13266`), and the description above has been rewritten to match.

**Rebased onto `main` (`da00543`).** The only conflict was the multi-requirement block, where
`d9b378598` ("support positional choices") had meanwhile added its own `p@N@` rule generation.
Resolved in favour of this branch's version, keeping main's conventions (`log.debug` calls,
`shlex.join`, quoting style, the template with the current header URL). Net diff is now 31 lines
inside `complete_tcsh`.

**Two fixes on top of the original version, both found by driving a real tcsh:**

1. `$COMMAND_LINE` must not be quoted in the padded `set cmd=(...)`. The original
   `set cmd=("$COMMAND_LINE" "" "" ...)` collapses the command line into `$cmd[1]`, leaving
   `$cmd[2]` empty forever, so *every* subcommand-level completion silently produced nothing —
   worse than before the patch. Now `set cmd=($COMMAND_LINE "" "" ...)`, which keeps the word
   splitting and still prevents out-of-range indexing.

2. The `c/--opt=/` rules were dead as originally ordered. tcsh stops at the first rule whose
   pattern matches, and the generic `'c/--/(...)'` line came first, so `--opt=<TAB>` was matched
   there, completed `opt=…` against the long-option names, found nothing, and gave up before ever
   reaching the specific rule. They are now emitted ahead of it via a separate `eq_specials` list.
   `--<TAB>` is unaffected, since `--` does not match `c/--opt=/`.

Both are covered by the before/after table in the description. `pre-commit` (bandit, flake8,
pyupgrade, mypy, yapf, isort) is clean.

shtab/__init__.py
```diff
@@ -765,13 +769,27 @@ def recurse_parser(cparser, positional_idx, requirements=None):
...
+                            # nested backticks crash tcsh's parser, use `eval` instead
+                            nlist.append(f"if ( {condition} ) eval {complete_fn.strip('`')}")
+                        else:
+                            nlist.append(f"if ( {condition} ) {complete_fn}")
+            if nlist:
+                nlist_str = '; '.join(nlist)
+                # pad $cmd so indexing it never runs out of range. $COMMAND_LINE must stay
+                # unquoted: quoting it stops csh word splitting, so $cmd[1] would be the whole
+                # command line and $cmd[2] always empty (no subcommand would ever complete).
+                padding = '"" "" "" "" "" "" "" "" ""'
```
casperdcl:
```suggestion
                # pad $cmd so indexing it never runs out of range.
                # $COMMAND_LINE must stay unquoted to allow csh word splitting.
                padding = ' '.join(['""'] * 9)
```

shtab/__init__.py
```diff
@@ -780,20 +798,25 @@ def recurse_parser(cparser, positional_idx, requirements=None):
             # Don't add a space after completing "--" from "-"
             optionals_single = ('-', '-')
 
+    # removes duplicates from list, preserves order
```
casperdcl:
```suggestion
```

casperdcl (approved):
Thanks!

PR #174, opened 2024-07-23
Supports fish shell

soraxas:
Closes #134

PR #216, opened 2026-04-07
feat: Add support for fish shell

### LINKED ISSUE #134: fish shell support

OdysseyOfTheDragons:
Closes #195.
Closes #174.
Closes #134.

This implementation uses `not __fish_seen_subcommand_from` to support subcommands and nested commands. Contrary to #195, this isn't limited to 3 levels deep. One limitation from this implementation is if a tool has subcommands `a b` and `b a`, then the sub-subcommands of both will be mixed.

### DIFF
--- a/examples/customcomplete.py
+++ b/examples/customcomplete.py
@@ -10,7 +10,7 @@ import shtab  # for completion magic
 
 TXT_FILE = {
     "bash": "_shtab_greeter_compgen_TXTFiles", "zsh": "_files -g '(*.txt|*.TXT)'",
-    "tcsh": "f:*.txt"}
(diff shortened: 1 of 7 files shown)


casperdcl (approved):
lgtm... rebased, fixed errors, added equivalent of #196

examples/customcomplete.py
casperdcl, line 1: @OdysseyOfTheDragons looks like all positional args aren't used

```sh
PYTHONPATH=. python examples/pathcomplete.py -s fish
```

is missing `input_txt` and `suffix` positional arg completions

casperdcl: fixed in d9b3785 (with the same mixed ordering issue as subcommands)

PR #225, opened 2026-07-31
zsh: restore escape_zsh for help text

### LINKED ISSUE #224: zsh: shlex.quote() in help breaks generated script — any apostrophe corrupts quoting (regression in 1.8.2)
Regression from #196, commit 15902075f32f591abd33e84c7f959b717713d378 ("Use shlex.quot() to do escaping"). Present in 1.8.2, 1.9.0 and still on `main` (61a0e60); 1.8.1 is fine.

Found via borgbackup CI (https://github.com/borgbackup/borg/issues/9982), but it reproduces standalone.

## The problem

`shlex.quote()` returns a *self-quoting standalone shell word*: it wraps the text in single quotes and encodes an embedded `'` as `'"'"'`. That idiom is only valid at top level. But the help text is interpolated into a **double-quoted** zsh word — the spec template is unchanged:

```python
'{nargs}{options}"[{help}]:{dest}:{pattern}"'
```

Dropping `'"'"'` inside `"…"` flips the quote parity, so every quote after it lands one position off and the rest of the spec ends up unquoted. Net effect: **a single apostrophe anywhere in any help string corrupts the whole generated zsh script.**

## Reproducer

```python
from argparse import ArgumentParser
import shtab

p = ArgumentParser(prog="demo")
p.add_argument("--x", help="don't do this")
p.add_argument("--sort-by", help="e.g. '>size_added,path'")
print(shtab.complete(p, shell="zsh"))
```

```console
$ python repro.py > repro.zsh
$ zsh -n repro.zsh
repro.zsh:24: parse error near `>'
```

Generated:

```zsh
15:  "--x['don'"'"'t do this']:x:"
16:  "--sort-by['e.g. '"'"'>size_added,path'"'"'']:sort_by:"
...
24:  local context state line curcontext="$curcontext" one_or_more='(*)' remainder='(-)*:' default='*::: :->demo'
```

How zsh tokenizes line 16:

| segment | quoting state |
| --- | --- |
| `--sort-by['e.g. '` | double-quoted |
| `"` | single-quoted |
| `>size_added,path` | **bare** — `>` is a redirection operator |
| `"` | single-quoted |
| `']:sort_by:` | double-quoted |

Inside an `array=( … )` assignment a redirection is a syntax error, hence `parse error near '>'`. (At top level it silently *redirects*: `print -r -- <that word>` creates a file named `size_added,path`.)

The `>` is incidental. A plain apostrophe alone (line 15) leaves an **unterminated** quote that cascades down the file until it reaches shtab's own `default='*::: :->{name}'` on line 24, whose `->` then goes bare — which is why the error is reported there and not at line 15.

Most of the damage is silent rather than fatal: in borg's generated script 32 specs contain the broken idiom, and only the first one to expose a `>` produces a parse error; the rest are just mis-quoted.

`escape_zsh()` (removed in that commit) backslash-escaped specials, which was safe inside `"…"`.

## Suggested fix

`shlex.quote` isn't the right tool for these fields — they aren't standalone shell words, they're fragments inside a composite `_arguments` spec that carries its own `[`, `]`, `:` structure, and `pattern` must stay unquoted so zsh evaluates it. Restoring a double-quote-safe escaper (i.e. `escape_zsh`, or at minimum escaping `"` `$` `` ` `` `\` plus `]` and `:`) for `help` looks like the minimal fix. Affected call sites on `main`:

- `format_optional` — `shtab/__init__.py:485`
- `format_positional` — `shtab/__init__.py:498`
- `command_list` — `commands = "\n    ".join(f'{quote(cmd)}:{quote(opt["help"])}' …)`

A regression test asserting `zsh -n` (or at least balanced quoting) over a script generated from a help string containing `'` would catch this class of bug — the current tests only compare generated text.

## Note on the bash side of #196

While bisecting I also hit `shlex.join()` being called on a `str` in the bash path, shredding `_COMPGEN` into space-separated characters:

```console
$ python -c "import shlex; print(shlex.join('_shtab_compgen_files'))"
_ s h t a b _ c o m p g e n _ f i l e s
```

That one is **already fixed** on `main` by e131b910aebd6d86395baf758080b1bcb33d19f4 — mentioning it only because it's still broken in released 1.9.0, so a release would be welcome once the zsh issue above is sorted.

## Versions

zsh 5.9, Python 3.11 and 3.14, macOS 15.6 — nothing platform-specific, `zsh -n` is pure parsing.

Happy to send a PR if you'd like.

ThomasWaldmann:
Fixes #224 — the zsh half of the #196 regression (the bash half is already fixed by e131b910aebd6d86395baf758080b1bcb33d19f4).

## The problem

`shlex.quote` returns a self-contained shell *word*: single-quoted, with an embedded `'` written as `'"'"'`. That is only valid at the top level. But `format_optional` and `format_positional` interpolate the result **inside** `"…"`:

```python
'{nargs}{options}"[{help}]:{dest}:{pattern}"'
```

There those quotes are literal, so the parity flips and the remainder of the spec goes unquoted. Two symptoms:

**Every** description with a space gained stray literal `'`, because `shlex.quote` quotes on spaces alone:

```zsh
# before (what zsh actually assigns)
(- : *)--help['show this help message and exit']
# after
(- : *)--help[show this help message and exit]
```

And a single apostrophe anywhere left an *unterminated* quote that cascaded down the file until it reached shtab's own `default='*::: :->{name}'`, whose `->` then parsed as a redirection:

```console
$ borg completion zsh | zsh -n
(stdin):715: parse error near `>'
```

## The fix

Restore `escape_zsh()` for the two call sites that interpolate into `"…"`. Backslash-escaping is safe there, and `_arguments` strips the backslashes when it parses the spec — which is why this worked from 1.8.0 through 1.8.1.

`command_list` and the bash/fish call sites use `quote()` in bare-word context, where it *is* correct — left alone.

## Test

`test_zsh_help_quoting` checks generated specs against real zsh rather than comparing text: `zsh -n` for syntax, then the values zsh assigns to the spec array, asserting the escaping introduces no *extra* quotes. It fails on all five help variants without the source change — including plain `"plain help"`, which shows the breakage wasn't limited to apostrophes:

```
E   assert 2 == 0
E    +  where 2 = "--opt['plain help']:opt:".count("'")
E    +  and   0 = 'plain help'.count("'")
```

Skipped via `shutil.which("zsh")` when zsh is unavailable.

## Verification

- `pytest`: 83 passed, coverage 87% (the one failure here, `test_path_completion_after_redirection`, is pre-existing on `main` — local bash 3.2 on macOS).
- flake8, isort, yapf, mypy: clean.
- End-to-end on borgbackup (https://github.com/borgbackup/borg/issues/9982): `borg completion zsh | zsh -n` passes again, and borg's completion test suite returns to its exact 1.8.1 result.

## Known pre-existing issue, deliberately not touched

`format_positional` puts help in `_arguments`' `message` field, which is `:`-delimited. A `:` in help is escaped to `\:` and does reach `_arguments` correctly, but I did not audit that path further — it behaves identically to 1.8.1, so it's out of scope for this regression fix. Happy to look at it separately if you'd like.

# Please review!

Note: fix is AI generated by Claude Opus 5. I did not review it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,6 +17,7 @@ jobs:
     - uses: actions/setup-python@v7
       with:
         python-version: ${{ matrix.python }}
+    - run: sudo apt-get update && sudo apt-get install -y zsh  # for generated-script tests
     - run: pip install -e .[dev]
     - run: pytest --cov-report=xml --cov-report=term
     - uses: codecov/codecov-action@v7
--- a/shtab/__init__.py
+++ b/shtab/__init__.py
@@ -459,10 +459,7 @@ def escape_zsh(string):
     """
     Backslash-escape for interpolation into a double-quoted `_arguments` spec.
 
-    NB: `shlex.quote` is unusable here. It returns a self-contained shell *word*
-    (single-quoted, with `'` written as `'"'"'`), which is only valid at the top
-    level. The specs below embed the result inside `"..."`, where those quotes are
-    literal and unbalance the surrounding string.
+    NOTE: cannot use `shlex.quote` (a single-quoted word only valid at top level).
     """
     # excessive but safe
     return re.sub(r"([^\w\s.,()-])", r"\\\1", str(string))
@@ -505,7 +502,7 @@ def complete_zsh(parser, root_prefix=None, preamble="", choice_functions=None):
                             if is_opt_end(opt) else '"*"' if is_opt_multiline(opt) else ""),
                      options=("{{{}}}".format(",".join(opt.option_strings)) if len(
                          opt.option_strings) > 1 else '"{}"'.format("".join(opt.option_strings))),
-                     help=escape_zsh(get_help(opt) if opt.help else ""),
+                     help=escape_zsh(get_help(opt)) if opt.help else "",
                      dest=opt.dest,
                      pattern=complete2pattern(opt.complete, "zsh", choice_type2fn) if hasattr(
                          opt, "complete") else choices2pattern(opt.choices) if opt.choices else "",
--- a/shtab/__init__.py
+++ b/shtab/__init__.py
@@ -455,6 +455,19 @@ complete -o filenames -F ${root_prefix} ${prog}""").safe_substitute(
     )
 
 
+def escape_zsh(string):
+    """
+    Backslash-escape for interpolation into a double-quoted `_arguments` spec.
+
+    NB: `shlex.quote` is unusable here. It returns a self-contained shell *word*
+    (single-quoted, with `'` written as `'"'"'`), which is only valid at the top
+    level. The specs below embed the result inside `"..."`, where those quotes are
+    literal and unbalance the surrounding string.
+    """
+    # excessive but safe
+    return re.sub(r"([^\w\s.,()-])", r"\\\1", str(string))
+
+
 @mark_completer("zsh")
 def complete_zsh(parser, root_prefix=None, preamble="", choice_functions=None):
     """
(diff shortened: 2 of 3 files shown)


casperdcl (approved):
Thanks!

Issue #231, opened 2026-08-01
fish: %(default)s not expanded in descriptions; subcommands lack descriptions when only help= is set


ThomasWaldmann:
Two smaller description-handling gaps in the fish backend, compared to zsh:

1. **`%(default)s` placeholders are not expanded.** argparse help strings commonly embed `%(default)s`/`%(prog)s` etc.; the fish backend passes `action.help` through verbatim:

   ```
   complete -c myprog ... -l retries -d 'number of retries (default: %(default)s)'
   ```

   The zsh backend expands this correctly (`--retries[number of retries (default\: 3)]`). Expansion is essentially `action.help % {**vars(action), "prog": ...}` (what argparse's `HelpFormatter._expand_help` does).

2. **Subcommand completions get no description when only `help=` is set.** `complete_fish` takes the description from `subparser.description`:

   ```python
   if subparser.description:
       desc = subparser.description.strip().split("\n")[0]
   ```

   The common argparse pattern is `sub.add_parser("create", help="create something")`, which leaves `description` as None — so with the repro parser, the generated `complete -c myprog -n "__fish_use_subcommand" -a create` has no `-d` at all. Falling back to the pseudo-action help (`_choices_actions`) would cover the common case.

Observed with shtab 1.9.2, fish 4.8.1, Python 3.11 on macOS.

Repro parser (`repro.py`, prints the completion script):

```python
import argparse, shtab

p = argparse.ArgumentParser(prog="myprog")
p.add_argument("--repo", "-r", help="repository to use")
sub = p.add_subparsers(dest="cmd")
c = sub.add_parser("create", help="create something")
c.add_argument("--exclude-from", help="read exclude patterns from FILE").complete = shtab.FILE
c.add_argument("--retries", type=int, default=3, help="number of retries (default: %(default)s)")
c.add_argument("name", choices=["alpha", "beta"], help="name of the thing to create")
c.add_argument("paths", nargs="*", help="paths to add")
d = sub.add_parser("delete", help="delete something")
d.add_argument("--force", action="store_true", help="force deletion")
d.add_argument("name", help="name of the thing to delete")
lp = sub.add_parser("list", help="list things")
lp.add_argument("--short", action="store_true", help="short output")

print(shtab.complete(p, "fish"))
```

Tested via `python repro.py > myprog.fish`, then in fish: `source myprog.fish; complete -C'myprog ...'`.

PR #232, opened 2026-08-01
fish: token-aware conditions, positional slots, file completion

### LINKED ISSUE #227: fish: no file completion; options marked shtab.FILE complete the wrong candidates
The fish backend cannot complete files at all, and options declared with `.complete = shtab.FILE` complete the wrong thing.

Two root causes in `complete_fish`:

1. The template globally disables file completion with `complete -c ${prog} -f`, and `CHOICE_FUNCTIONS["file"]["fish"]` is the empty string, so `shtab.FILE` expands to nothing — there is no file completion story for fish.
2. Options that take a value get no `-r`/`--require-parameter` flag (only options with `choices`/`.complete` get `-rka`, and for `FILE` even that collapses to nothing). Without `-r`, fish does not know the next token belongs to the option, so it completes that position as if it were a positional argument.

Behavior with the repro below:

```
> complete -C'myprog create alpha my'          # paths positional
                                               # (nothing — expected: file completion)
> complete -C'myprog create --exclude-from '   # explicitly marked shtab.FILE
alpha  (name of the thing to create)
beta   (name of the thing to create)           # (expected: file completion)
```

The second case is actively wrong: the argument of `--exclude-from` is offered the choices of the `name` positional.

For comparison, the bash and zsh backends complete files in both places.

Suggested direction: emit `-r` for every option that takes a value; map `FILE` to real file completion (e.g. drop the global `-f` in favor of per-entry `-f`, or use `-rF`), analogous to how `DIRECTORY` already maps to `(__fish_complete_directories)`.

Observed with shtab 1.9.2, fish 4.8.1, Python 3.11 on macOS.

Repro parser (`repro.py`, prints the completion script):

```python
import argparse, shtab

p = argparse.ArgumentParser(prog="myprog")
p.add_argument("--repo", "-r", help="repository to use")
sub = p.add_subparsers(dest="cmd")
c = sub.add_parser("create", help="create something")
c.add_argument("--exclude-from", help="read exclude patterns from FILE").complete = shtab.FILE
c.add_argument("--retries", type=int, default=3, help="number of retries (default: %(default)s)")
c.add_argument("name", choices=["alpha", "beta"], help="name of the thing to create")
c.add_argument("paths", nargs="*", help="paths to add")
d = sub.add_parser("delete", help="delete something")
d.add_argument("--force", action="store_true", help="force deletion")
d.add_argument("name", help="name of the thing to delete")
lp = sub.add_parser("list", help="list things")
lp.add_argument("--short", action="store_true", help="short output")

print(shtab.complete(p, "fish"))
```

Tested via `python repro.py > myprog.fish`, then in fish: `source myprog.fish; complete -C'myprog ...'`.

### LINKED ISSUE #228: fish: no subcommand completion after a global option that takes a value
The fish backend guards all top-level completions (subcommand names and global options) with `-n "__fish_use_subcommand"`. That fish function returns false as soon as *any* non-option token is on the command line — including the **value** of a global option. So the very common pattern "global option first, then subcommand" gets no completions at all:

```
> complete -C'myprog --repo x '
                                   # (nothing — expected: create, delete, list)
```

The same happens for `-r x`, or any other global option that consumes a value.

Fixing this needs knowledge of which option strings take a value, which the generator has (from the parser's actions). One approach: emit a script-level list of value-taking option strings and use a custom condition function that skips options *and their values* when deciding whether a subcommand has been given yet, instead of `__fish_use_subcommand`.

Observed with shtab 1.9.2, fish 4.8.1, Python 3.11 on macOS.

Repro parser (`repro.py`, prints the completion script):

```python
import argparse, shtab

p = argparse.ArgumentParser(prog="myprog")
p.add_argument("--repo", "-r", help="repository to use")
sub = p.add_subparsers(dest="cmd")
c = sub.add_parser("create", help="create something")
c.add_argument("--exclude-from", help="read exclude patterns from FILE").complete = shtab.FILE
c.add_argument("--retries", type=int, default=3, help="number of retries (default: %(default)s)")
c.add_argument("name", choices=["alpha", "beta"], help="name of the thing to create")
c.add_argument("paths", nargs="*", help="paths to add")
d = sub.add_parser("delete", help="delete something")
d.add_argument("--force", action="store_true", help="force deletion")
d.add_argument("name", help="name of the thing to delete")
lp = sub.add_parser("list", help="list things")
lp.add_argument("--short", action="store_true", help="short output")

print(shtab.complete(p, "fish"))
```

Tested via `python repro.py > myprog.fish`, then in fish: `source myprog.fish; complete -C'myprog ...'`.

### LINKED ISSUE #229: fish: completions vanish when an argument value equals a subcommand name
The fish backend builds per-subcommand conditions from `__fish_seen_subcommand_from`, e.g. for the repro parser:

```
complete -c myprog -n "__fish_seen_subcommand_from delete; and not __fish_seen_subcommand_from list create" -l force -d 'force deletion'
```

`__fish_seen_subcommand_from` matches **any** token on the command line, not just the token in subcommand position. So as soon as an argument *value* happens to equal a sibling subcommand name, the conditions of *both* subcommands go false and all completions vanish:

```
> complete -C'myprog delete list --'
                                   # (nothing — expected: --force, --help)
```

Here `list` is the value of `delete`'s `name` positional, but the condition machinery reads it as "the list subcommand was seen": `delete`'s options are suppressed by `not __fish_seen_subcommand_from list ...`, and `list`'s options are suppressed by `not ... delete`, so nothing is offered.

This is not exotic: in borgbackup for example, an archive named `list`, `check` or `prune` (all plausible archive names) makes completion for the rest of the command line go dark.

A robust fix needs a condition that determines the *actual* subcommand from token position (skipping options and their values — related to the value-skipping needed for #228), rather than "token seen anywhere".

Observed with shtab 1.9.2, fish 4.8.1, Python 3.11 on macOS.

Repro parser (`repro.py`, prints the completion script):

```python
import argparse, shtab

p = argparse.ArgumentParser(prog="myprog")
p.add_argument("--repo", "-r", help="repository to use")
sub = p.add_subparsers(dest="cmd")
c = sub.add_parser("create", help="create something")
c.add_argument("--exclude-from", help="read exclude patterns from FILE").complete = shtab.FILE
c.add_argument("--retries", type=int, default=3, help="number of retries (default: %(default)s)")
c.add_argument("name", choices=["alpha", "beta"], help="name of the thing to create")
c.add_argument("paths", nargs="*", help="paths to add")
d = sub.add_parser("delete", help="delete something")
d.add_argument("--force", action="store_true", help="force deletion")
d.add_argument("name", help="name of the thing to delete")
lp = sub.add_parser("list", help="list things")
lp.add_argument("--short", action="store_true", help="short output")

print(shtab.complete(p, "fish"))
```

Tested via `python repro.py > myprog.fish`, then in fish: `source myprog.fish; complete -C'myprog ...'`.

### LINKED ISSUE #230: fish: positional argument order ignored (earlier positionals' candidates re-offered at later slots)
`complete_fish` emits the completions of *all* positional arguments of a (sub)command under the same condition — the code notes this itself:

```python
# NOTE/caveat: positional argument order is ignored
```

Consequences with the repro parser (`create` takes `name {alpha,beta}` followed by `paths ...`):

```
> complete -C'myprog create alpha alp'
alpha  (name of the thing to create)   # (name choices re-offered at the paths position)
```

The `name` choices are offered again for every later positional slot (and, combined with #227, the `paths` slot offers nothing that *should* be there). Similarly, once a single-valued positional has been given, its candidates keep being offered.

This makes completion misleading for any command with more than one positional (`prog ARCHIVE PATH...`, `prog SRC DST`, ...).

Suggested direction: count the positional tokens already present (skipping options and their values, cf. #228/#229) and guard each positional's completions with a positional-index condition (`-eq N` for single-valued, `-ge N` for `nargs='*'/'+'/...`). This is the approach borgbackup now uses in its own fish completion generator, where it works well.

Observed with shtab 1.9.2, fish 4.8.1, Python 3.11 on macOS.

Repro parser (`repro.py`, prints the completion script):

```python
import argparse, shtab

p = argparse.ArgumentParser(prog="myprog")
p.add_argument("--repo", "-r", help="repository to use")
sub = p.add_subparsers(dest="cmd")
c = sub.add_parser("create", help="create something")
c.add_argument("--exclude-from", help="read exclude patterns from FILE").complete = shtab.FILE
c.add_argument("--retries", type=int, default=3, help="number of retries (default: %(default)s)")
c.add_argument("name", choices=["alpha", "beta"], help="name of the thing to create")
c.add_argument("paths", nargs="*", help="paths to add")
d = sub.add_parser("delete", help="delete something")
d.add_argument("--force", action="store_true", help="force deletion")
d.add_argument("name", help="name of the thing to delete")
lp = sub.add_parser("list", help="list things")
lp.add_argument("--short", action="store_true", help="short output")

print(shtab.complete(p, "fish"))
```

Tested via `python repro.py > myprog.fish`, then in fish: `source myprog.fish; complete -C'myprog ...'`.

ThomasWaldmann:
Fixes #227, fixes #228, fixes #229, fixes #230.

Reworks the fish backend to parse the command line properly instead of relying on `__fish_use_subcommand` / `__fish_seen_subcommand_from`:

- The generated script contains a small, prefix-namespaced scan function (`{prefix}_scan`) that walks `commandline -opc` once and determines the *actual* (sub)command path (e.g. `sub subsub`) and the number of positional arguments given after it — skipping options and, using a generated list of value-taking option strings, their values. All completions are guarded by conditions on that state (`{prefix}_using <path>`, `test ${prefix}_npos -eq N`).
- Consequences:
  - `prog --global-opt value <TAB>` completes subcommands again (#228: `__fish_use_subcommand` treated the option value as "subcommand already given").
  - argument values that happen to equal a sibling command name no longer make all completions vanish (#229: `__fish_seen_subcommand_from` matches any token).
  - each positional argument's completions are restricted to its slot(s) — `-eq N` for single-valued positionals, `-ge N` once `nargs='*'/'+'/...` starts — instead of "positional argument order is ignored" (#230).
- File completion works now (#227):
  - the global `complete -c prog -f` is gone; flag options get `-f` individually, positional-candidate entries get `-f` per entry.
  - options that take a value always get `-r`, so fish knows a value token follows (previously an option's value slot was completed like a positional).
  - value-taking options and positionals *without* completion info fall back to fish's default file completion, matching what the bash/zsh backends do.
  - `Choice` placeholders (`shtab.Optional.FILE` etc.) are honored instead of being emitted as literal candidate strings.
- `root_prefix` is no longer ignored for fish (it namespaces the generated functions/variables, as for bash/zsh).

New functional tests drive fish's own `complete -C` against generated scripts (skipped when fish is unavailable; fish is present on the CI runners' apt/brew? if not, they simply skip — the static `test_prog_scripts` expectations still cover the generated text). Existing suite passes unchanged apart from the updated fish expectations in `test_prog_scripts`.

Also verified against a large real-world parser (borgbackup, ~56 subcommands incl. two-level `borg key export` style groups and custom dynamic completions via `.complete` + preamble): subcommand, option, option-value, positional-slot and dynamic completions all behave correctly under `complete -C` with fish 4.8.1.

Known limitations (unchanged or best-effort, noted for review):
- options with `nargs` consuming more than one value token only have their first value skipped by the scan;
- a subcommand is only recognized at the first positional slot (parsers that put a regular positional *before* `add_subparsers` are not supported — same as before);
- single-dash negative-number arguments are treated as options (as everywhere else).

Context: found while replacing hand-written fish completions in borgbackup (borgbackup/borg#9990).

Created by Claude Fable 5, please review carefully.

### DIFF
--- a/shtab/__init__.py
+++ b/shtab/__init__.py
@@ -25,10 +25,11 @@ log = logging.getLogger(__name__)
 SUPPORTED_SHELLS: list[str] = []
 _SUPPORTED_COMPLETERS: dict[str, Callable] = {}
 CHOICE_FUNCTIONS: dict[str, dict[str, str]] = {
-    "file": {"bash": "_shtab_compgen_files", "zsh": "_files", "tcsh": "f", "fish": ""},
-    "directory": {
-        "bash": "_shtab_compgen_dirs", "zsh": "_files -/", "tcsh": "d",
-        "fish": "(__fish_complete_directories)"}}
+    "file": {
+        "bash": "_shtab_compgen_files", "zsh": "_files", "tcsh": "f",
+        "fish": "(__fish_complete_path)"}, "directory": {
+            "bash": "_shtab_compgen_dirs", "zsh": "_files -/", "tcsh": "d",
+            "fish": "(__fish_complete_directories)"}}
 FILE = CHOICE_FUNCTIONS["file"]
 DIRECTORY = DIR = CHOICE_FUNCTIONS["directory"]
 FLAG_OPTION = (
@@ -823,45 +824,49 @@ def complete_fish(parser, root_prefix=None, preamble="", choice_functions=None):
     """
     Return fish syntax autocompletion script.
 
-    root_prefix:
-      ignored (fish has no support for functions)
-
-    See `complete` for other arguments.
+    See `complete` for arguments.
     """
     prog = parser.prog
-    completions = []
+    prefix = wordify(f"_shtab_{root_prefix or prog}")
+    completions: list = []
+    # all (sub)command paths, e.g. ["sub", "sub subsub"]
+    commands: list = []
+    # option strings which consume a following value token
+    opts_with_value: set = set()
 
     choice_type2fn = {k: v["fish"] for k, v in CHOICE_FUNCTIONS.items()}
     if choice_functions:
         choice_type2fn.update(choice_functions)
 
-    def start_output(path, same_level):
-        output = ["complete", "-c", prog]
-        if len(path) > 0:
-            cond = '; and '.join(f"__fish_seen_subcommand_from {cmd}" for cmd in path)
-            if same_level != "":
-                cond += f"; and not __fish_seen_subcommand_from {same_level}"
-            output.append(f'-n "{cond}"')
-        else:
-            output.append('-n "__fish_use_subcommand"')
-        return output
+    def is_flag(opt):
+        return isinstance(opt, FLAG_OPTION) or opt.nargs == 0
 
-    def get_specials(arg):
-        if arg.choices:
-            yield f'-rka "{join(map(str, arg.choices))}"'
-        elif hasattr(arg, 'complete'):
-            complete_fn = complete2pattern(arg.complete, 'fish', choice_type2fn)
-            if complete_fn:
-                yield f'-rka "{complete_fn}"'
+    def get_pattern(arg):
+        """`Choice`/`.complete` completion pattern, or None if there is none."""
+        if arg.choices and isinstance(next(iter(arg.choices)), Choice):
+            return choice_type2fn[next(iter(arg.choices)).type]
+        if hasattr(arg, "complete"):
+            return complete2pattern(arg.complete, "fish", choice_type2fn)
+        return None
 
-    def recurse_parser(cparser: ArgumentParser, path: list[str], same_level: str):
+    def pos_condition(index, width, open_ended):
+        """Condition suffix restricting a completion to the given positional slot(s)."""
+        npos = f"${prefix}_npos"
+        if open_ended or width is None:
+            return f"; and test {npos} -ge {index}"
+        if width == 1:
+            return f"; and test {npos} -eq {index}"
+        return f"; and test {npos} -ge {index}; and test {npos} -le {index + width - 1}"
+
+    def start_output(path, pos_test=""):
+        """`complete` command start, with a condition matching the (sub)command `path`."""
+        cond = " ".join([f"{prefix}_using"] + [quote(cmd) for cmd in path]) + pos_test
+        return ["complete", "-c", prog, f"-n {quote(cond)}"]
+
+    def recurse_parser(cparser: ArgumentParser, path: list[str]):
         """
         path:
           the list of subcommands that led to current
-
-        same_level:
-          a space separated list of other subcommands available on the same level
-          must not contain the current subcommand
         """
         log_prefix = "| " * len(path)
         log.debug("%sParser @ %d", log_prefix, len(path))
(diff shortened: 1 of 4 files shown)


casperdcl (changes requested):
- ~~missing `help` is intended to mean a hidden (undocumented, beta/hidden) CLI argument, so shouldn't fallback to `description` nor have any completions generated~~ moved discussion to #233
- global `complete -c prog -f` is desired (for all shells). Users should explicitly use `shtab.FILE` to enable that sort of completion

Feel free to tweak the [FAQs](https://tqdm.github.io/shtab/#faqs) (source: `docs/index.md`) to make the above clearer

ThomasWaldmann:
Thanks for the review — both points are applied, here (`f4ee8c0`) and in the stacked #233 (`540a97d`).

### Global `complete -c prog -f`

Restored, and path completion is now strictly opt-in:

- the per-entry `-f`s are gone again (flags, subcommand entries, positionals);
- `CHOICE_FUNCTIONS["file"]["fish"]` is now `(__fish_complete_path)` instead of `""`, so `shtab.FILE` actually completes files in fish — analogous to `DIRECTORY`'s `(__fish_complete_directories)`. That is what remains of #227: with the global `-f` in place *and* an empty pattern, `shtab.FILE` expanded to nothing, so explicitly asking for file completion had no effect;
- arguments with no completion info (positionals, and the values of value-taking options) now complete nothing at all;
- value-taking options get `-x` (i.e. `-r -f`) rather than `-r`.

That last one is what actually makes the global `-f` stick. `-f` at the command level does not survive an entry that takes a parameter: with plain `-r`, fish file-completes the value anyway, and `-rka "…"` — what `main` emits today — appends the file list to the candidates. So on `main`, with `examples/pathcomplete.py`:

```console
> complete -C'pathcomplete --dir a'
adir/  (Directory)
alpha.txt              # a file, from a DIRECTORY-marked option

> complete -C'pathcomplete -s '
bash  zsh  tcsh  fish  # the choices...
adir/  alpha.txt  beta.txt  …    # ...plus the entire cwd
```

`-xka` fixes both (fish 4.8.1).

### Missing `help` = hidden

#233 no longer falls back: the subcommand description now comes from the `help` passed to `add_parser`, never from `description`. Subparsers without `help` were already dropped via `get_public_subcommands` (argparse only records a `_ChoicesPseudoAction` when `help=` is passed to `add_parser`), and there is now a test asserting such a subparser produces no completions at all.

### FAQs

Two bullets in `docs/index.md`:

> - Make sure that all arguments have `help` messages (…). Arguments (and `add_parser()` subcommands) without `help` are treated as hidden - i.e. deliberately undocumented (e.g. beta/internal) - so no completions are generated for them.
> - Make sure that arguments which should complete file/directory paths are marked as such (`parser.add_argument('positional').complete = shtab.FILE`, see Usage). Path completion is *never* enabled by default - not even for arguments without any other completion information.

Reword freely if that is not quite the intent.

### Two things I would like your call on

**1. `help` vs `description` for subcommand descriptions.** fish now shows the `add_parser(help=…)` one-liner; `complete_zsh` still shows `subparser.description` ([`__init__.py#L565`](https://github.com/tqdm/shtab/blob/225e3ff143c90ce8476a541436b96a239ccca65d/shtab/__init__.py#L565)). So the two backends now disagree for a parser that sets both. My reading of "shouldn't fallback to `description`" is that `help` is the intended source and `description` is the long prose for `prog sub --help` — but if you would rather keep `description` as the source everywhere, say so and I will flip fish back and drop that part of #231. Either way, should I align zsh in the same PR?

**2. How far does "nor have any completions generated" go?** Right now only `help == SUPPRESS` suppresses an argument in *any* backend; an option or positional with `help=None` is still completed, just without a description. Subcommands are the only case where a missing `help` already means hidden, and that falls out of argparse rather than shtab. If `help=None` should mean hidden for options and positionals too, that is a cross-backend behaviour change — happy to do it, but I would rather send it as its own PR than fold it into the fish rework. Which would you prefer?

Also flagging, since it is not a fish bug: the new `diff` job fails on both PRs by design, because the generated fish scripts necessarily change.

<sub>These changes and this comment were prepared by Claude Opus 5 — please review carefully.</sub>

shtab/__init__.py
```diff
@@ -823,99 +824,179 @@ def complete_fish(parser, root_prefix=None, preamble="", choice_functions=None):
...
-            if complete_fn:
-                yield f'-rka "{complete_fn}"'
+    def is_flag(opt):
+        return isinstance(opt, FLAG_OPTION) or opt.nargs == 0
 
-    def recurse_parser(cparser: ArgumentParser, path: list[str], same_level: str):
+    def get_pattern(arg):
+        """`Choice`/`.complete` completion pattern, or None if there is none."""
+        if arg.choices and isinstance(next(iter(arg.choices)), Choice):
+            return choice_type2fn[next(iter(arg.choices)).type]
```
casperdcl: `shtab.Choice` is soft-deprecated. It's still in `bash/zsh` for backwards-compat, but shouldn't be included elsewhere (`tcsh`, `fish`)

casperdcl (changes requested):

tests/test_shtab.py
```diff
@@ -302,6 +301,107 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+        completion = shtab.complete(get_fish_test_parser(), shell="fish")
+
+    # positional marked `shtab.FILE` (`paths`, after the `name` slot)
+    assert "test_file.txt" in fish_candidates(completion, "test create alpha test_")
+    # value of an option marked `shtab.FILE` completes files, not the positional's choices
+    candidates = fish_candidates(completion, "test create --exclude-from ")
+    assert "test_file.txt" in candidates
+    assert "alpha" not in candidates
+    # arguments not marked `shtab.FILE` don't complete files (as in the other shells)
+    assert fish_candidates(completion, "test delete test_") == []
```
casperdcl: any idea why the CI fails here? Can't reproduce the failure locally

casperdcl: needed `fish>=4`

